PR #6, opened 2026-03-10
feat: plugin venv isolation

### LINKED ISSUE #5: [FEATURE]: Virtual Environment Isolation
### 🧭 Type of Feature

Please select the most appropriate category:

- [x] Enhancement to existing functionality
- [ ] Developer tooling or test improvement

### Virtual Environment Isolation

- [x] Design venv directory structure per plugin
- [x] Implement venv creation on plugin first load
- [x] Implement requirements.txt installation in venv
- [x] Execute plugin code within venv context
- [x] Cache venvs and detect requirements changes
- [x] Add configuration for venv base directory
- [ ] Add cleanup for orphaned venvs

---

### User Story 1

**As a:** plugin developer
**I want:** My plugin to run in an isolated virtual environment
**So that:** My plugin's dependencies don't conflict with other plugins

#### ✅ Acceptance Criteria
```gherkin
Scenario: Plugin runs in isolated venv
  Given I have a plugin with dependency "transformers==4.35.0"
  And another plugin requires "transformers==4.30.0"
  When both plugins are loaded
  Then each plugin should have its own virtual environment
  And both plugins should function correctly with their respective versions

Scenario: Plugin venv is created automatically
  Given I register a new native Python plugin
  When the plugin is loaded for the first time
  Then a virtual environment should be created for the plugin
  And the plugin's requirements.txt should be installed
  And the venv location should be logged

Scenario: Plugin venv is reused on restart
  Given a plugin has an existing virtual environment
  When the gateway restarts
  Then the existing venv should be reused
  And dependencies should not be reinstalled unless changed
```

---
### 📐 Design Sketch (optional)

Include a diagram, sketch, or flow (use [Mermaid](https://mermaid.js.org/) if desired):

---

---

tedhabeck:
### Summary

Closes: #5 

### Changes

This branch represents a significant architectural enhancement enabling plugins to run in isolated Python environments, improving security, dependency management, and plugin compatibility.

### Key Enhancements

1. Plugin Isolation via Virtual Environment (Primary Feature)

    #### New Core Modules:
    
    - cpex/framework/isolated/client.py  - Client for managing isolated plugin execution
    - cpex/framework/isolated/worker.py  - Worker process for executing plugins in isolation
    - cpex/framework/isolated/venv_comm.py - Communication layer between host and isolated environment

2. Venv Cache Support
Caching mechanism for virtual environments to improve performance
Reduces overhead of repeated venv creation

3. Serialization Support
Enhanced serialization capabilities for plugin data exchange
Supports communication between isolated environments

4. Configuration Support
New fixture: isolated_plugin.yaml for testing isolated plugin configurations
Updated loader to support isolated plugin mode

### Checks

- [x] `make lint` passes
- [x] `make test` passes

## Performance Optimization Complete ✓
Restructured the isolated venv plugin system to use a long-running worker process instead of forking a new subprocess for each invocation.

#### Before (forking new process each time):

prompt_pre_fetch: 0.147ms avg
prompt_post_fetch: 0.155ms avg
tool_pre_invoke: 0.143ms avg
tool_post_invoke: 0.149ms avg
Average: ~0.148ms per invocation
#### After (long-running worker process):

prompt_pre_fetch: 0.045ms avg
prompt_post_fetch: 0.047ms avg
tool_pre_invoke: 0.043ms avg
tool_post_invoke: 0.044ms avg
Average: ~0.045ms per invocation
Performance Gain: 3.3x faster (70% reduction in latency)

### Example Plugin directory structure with venv configured:

image


cpex/framework/isolated/client.py
```diff
@@ -0,0 +1,315 @@
...
+            logger.info(f"Removing existing venv at {venv_path}")
+            shutil.rmtree(venv_path_obj)
+
+        # Check Python version
+        python_version = sys.version_info
+        print(f"Current Python version: {python_version.major}.{python_version.minor}.{python_version.micro}")
+
+        # Create the EnvBuilder with common options
+        builder = venv.EnvBuilder(
+            system_site_packages=True,  # Don't include system site-packages
```
terylt: Is this meant to be set to true? The comment suggests that it shouldn't as it will share the system libraries.   Might be good to make this configurable?

cpex/framework/isolated/client.py
```diff
@@ -0,0 +1,315 @@
...
+            clear=False,  # Don't clear existing venv if it exists
+            symlinks=False,  # Use symlinks (recommended on Unix-like systems)
```
terylt: This is set to false but comment suggests True.

cpex/framework/isolated/client.py
```diff
@@ -0,0 +1,315 @@
...
+        """Initialize the plugin's venv environment with caching support."""
+        # ensure the config is validated
+        if not os.path.exists(self.plugin_path):
+            raise FileNotFoundError(f"plugin path not found: {self.plugin_path}")
+
+        venv_path = self.plugin_path / ".venv"
+        requirements_file = self.plugin_path / self.config.config["requirements_file"]
+
+        # Create venv with caching support
+        self.venv = await self.create_venv(venv_path=venv_path, requirements_file=requirements_file, use_cache=True)
```
terylt: create_venv doesn't return anything and is assigned to self.venv

terylt: Also I don't think self.venv is actually used anywhere.

cpex/framework/isolated/client.py
```diff
@@ -0,0 +1,315 @@
...
+    """IsolatedVenvPlugin class."""
+
+    def __init__(self, config: PluginConfig) -> None:
+        """Initialize the plugin's venv environment."""
+        super().__init__(config)
+        self.implementation = "Python"
+        self.comm = None
+        self.script_path: str = config.config["script_path"]
+        path = Path(self.config.config.get("script_path")).resolve()
+        class_root = self.config.config.get("class_name").split(".")[0]
```
terylt, line 45: If two plugins are in the same directory (package) and script path they may collide and overwrite each other's environment. How do we want to handle plugins in the same package? And should we validate against venv overwriting?

cpex/framework/isolated/client.py
```diff
@@ -0,0 +1,315 @@
...
+        }
+
+    async def invoke_hook(self, hook_type: str, payload: PluginPayload, context: PluginContext) -> PluginResult:
+        """Invoke a plugin in the context of the active venv (self.comm)"""
+        try:
+            # Validate and get result type
+            self._validate_hook_invocation(hook_type)
+
+            # Build and send task
+            task_data = self._build_hook_task(hook_type, payload, context)
```
terylt, line 321: Have you tried protobuf objects over stdio?   Just wondering if you'd get any speed up.  Serialization costs tend to be lower, and more structured with a proto file, but you do lose some human readability.  Might be interesting to try if it's not too hard. I wouldn't block the PR based on this but just a thought.

cpex/framework/isolated/venv_comm.py
```diff
@@ -0,0 +1,264 @@
...
+
+    def install_requirements(self, requirements_file: str) -> None:
+        """
+        Install Python requirements from a file in the target venv.
+        Args:
+            requirements_file (str): Path to the requirements file.
+        """
+        requirements_path = Path(requirements_file)
+        if requirements_path.exists():
+            rc = subprocess.check_call([self.python_executable, "-m", "pip", "install", "-r", requirements_file])
```
terylt: I think anything after this is dead code, as check_call() raises a subprocess.CalledProcessError if the command returns a non-zero exit status. so may want a try cache block.

cpex/framework/isolated/worker.py
```diff
@@ -0,0 +1,169 @@
+# -*- coding: utf-8 -*-
+"""Location: ./cpex/framework/isolated/server.py
```
terylt: file name is incorrect.

cpex/framework/isolated/worker.py
```diff
@@ -0,0 +1,169 @@
...
+    task_type = task_data.get("task_type")
+
+    if task_type == "info":
+        return {
+            "status": "success",
+            "environment": get_environment_info(),
+            "message": "Environment info retrieved successfully",
+        }
+    # This is essentially emulating the plugin loader's load and instantiate plugin
+    if task_type == "load_and_run_hook":
```
terylt, line 93: I'm concerned about the overhead and loss of state of reinitializing a plugin and loading it each time.   Can we  cache the Plugin during an initialization phase?   Could ship the individual plugin config over once in an initial phase.. load the plugin, and cache it and just use it if it's cached.

terylt: If we can separate this into a load and an invoke phase, this would be good.

cpex/framework/isolated/worker.py
```diff
@@ -0,0 +1,169 @@
...
+        # retrieve the context
+        context = task_data.get("context")
+        plugin_context = PluginContext(
+            state=context.get("state"), global_context=context.get("global_context"), metadata=context.get("metadata")
+        )
+        result = await executor.execute_plugin(
+            hook_ref, payload=task_data.get("payload"), local_context=plugin_context, violations_as_exceptions=False
+        )
+        return result
+
```
terylt, line 152: Returns none on unknown task types.  Should probably raise or return an error response.

cpex/framework/isolated/worker.py
```diff
@@ -0,0 +1,169 @@
...
+            "environment": get_environment_info(),
+            "message": "Environment info retrieved successfully",
+        }
+    # This is essentially emulating the plugin loader's load and instantiate plugin
+    if task_type == "load_and_run_hook":
+        # relative path from project root.
+        json_config = task_data.get("config")
+        config_raw = json.loads(json_config)
+        module_path: str = task_data.get("script_path")
+        sys.path.append(str(Path(module_path).resolve()))
```
terylt: sys.path.append with user-provided path                                                                                                                                
 sys.path.append(str(Path(module_path).resolve()))
This appends an arbitrary path to sys.path and then does importlib.import_module(mod_name). If module_path or class_name come from untrusted config, this is a code execution vector. 
  Consider validating against an allowlist like the root path.

cpex/framework/isolated/venv_comm.py
```diff
@@ -0,0 +1,264 @@
...
+        Args:
+            script_path (str): Path to the worker script
+        """
+        if self.running:
+            logger.warning("Worker process already running")
+            return
+
+        try:
+            # Start child process
+            self.process = subprocess.Popen(
```
terylt, line 82: Subprocess spawned without restricting environment                                                                                                               
  The worker process inherits the full parent environment. Consider passing a filtered env dict to Popen to limit what the isolated plugin can access (credentials, tokens, etc.).   This might require a bit of a bigger conversation.

cpex/framework/isolated/client.py
```diff
@@ -0,0 +1,315 @@
...
+        requirements_file = self.plugin_path / self.config.config["requirements_file"]
+
+        # Create venv with caching support
+        self.venv = await self.create_venv(venv_path=venv_path, requirements_file=requirements_file, use_cache=True)
+
+        self.comm = VenvProcessCommunicator(venv_path)
+
+        # Only install requirements if venv was newly created or cache was invalid
+        # Check if we need to install requirements
+        if not self._is_venv_cache_valid(venv_path, requirements_file):
```
terylt: client.py:219 — Double cache validation in initialize()                                                                                                                               
  create_venv already checks _is_venv_cache_valid internally. Then initialize() calls _is_venv_cache_valid again immediately after. There's a TOCTOU race between the check and the        
  install, and it's redundant work. Consider having create_venv return a boolean indicating whether the venv was freshly created.

cpex/framework/isolated/venv_comm.py
```diff
@@ -0,0 +1,264 @@
+# -*- coding: utf-8 -*-
+"""
+Location: ./cpex/framework/isolated/venv_comm.py
+Copyright 2025
+SPDX-License-Identifier: Apache-2.0
+Authors: Fred Araujo, Ted Habeck
+"""
+
+import json
```
terylt: This class uses both orjson and json packages in different spots. Probably just needs one.

cpex/framework/models.py
```diff
@@ -1567,3 +1612,302 @@ class PluginPayload(BaseModel):
...
+            ValueError: If the package name is invalid.
+        """
+        if pypi_package is not None and pypi_package != "":
+            # PyPI package names must contain only ASCII letters, numbers, hyphens, underscores, and periods
+            # They cannot start or end with hyphens or periods
+            if not pypi_package.strip():
+                raise ValueError("PyPI package name cannot be empty or whitespace")
+
+            # Check for valid characters
+            import re
```
terylt: Each of validate_pypi_package, validate_git_repository, validate_git_branch_tag_commit, and validate_version_constraint does import re at the function level. Move it to the module      
  top-level import.

cpex/framework/models.py
```diff
@@ -1295,6 +1298,48 @@ def check_config_and_external(self, info: ValidationInfo) -> Self:  # pylint: di
...
+            PluginConfig: A new PluginConfig instance with only data fields.
+        """
+        # Get the JSON-safe dictionary representation
+        safe_data = self.toJSON()
+
+        # Create a new PluginConfig instance from the safe data
+        # This will run validators again, but the resulting object will be clean
+        return orjson.dumps(safe_data).decode()
+
+    def toJSON(self) -> dict[str, Any]:
```
terylt: This method name violates PEP 8. Could be: to_json()

cpex/framework/isolated/client.py
```diff
@@ -0,0 +1,315 @@
...
+        Args:
+            venv_path: Path where the virtual environment should be created
+            requirements_file: Path to requirements file for cache validation
+            use_cache: Whether to use cached venv if available
+        """
+        venv_path_obj = Path(venv_path)
+
+        # Check if we can use cached venv
+        if use_cache and requirements_file and self._is_venv_cache_valid(venv_path, requirements_file):
+            logger.info(f"Using cached virtual environment at: {venv_path_obj.resolve()}")
```
terylt: Using f-strings in logging calls (e.g., logger.info(f"...")) instead of logger.info("...", arg). This evaluates the f-string even when the log level  is disabled.  This happens in multiple spots.

cpex/framework/isolated/client.py
```diff
@@ -0,0 +1,315 @@
...
+        try:
+            builder.create(venv_path)
+            print("✓ Virtual environment created successfully!")
+            print("\nTo activate the virtual environment:")
+            print(f"  source {venv_path}/bin/activate  # On Unix/macOS")
+            print(f"  {venv_path}\\Scripts\\activate  # On Windows")
+
+        except Exception as e:
+            print(f"✗ Error creating virtual environment: {e}")
+            raise e
```
terylt: can just be raise to preserve the original traceback

cpex/framework/models.py
```diff
@@ -14,7 +14,10 @@
 import os
 from enum import Enum, StrEnum
 from pathlib import Path
-from typing import Any, Generic, Optional, Self, TypeVar, Union
+from typing import Any, Generic, List, Optional, Self, TypeVar, Union
```
terylt, line 18: List is unnecessary in python 3.9+, can just use list[] directly

terylt (dismissed):
Overall, really nice work @tedhabeck.  I made some comments throughout the code for improvements.

I like:  
- The long-running worker process design (vs. fork-per-invocation) is a solid architectural choice for performance.
  - Venv caching via requirements hash is practical and well-implemented.                                                                                                                  
  - The request-id based response routing in VenvProcessCommunicator handles concurrent requests cleanly.                                                                                  
  - Graceful shutdown with fallback to kill() is well handled.                                                                                                                             
  - The PluginPackageInfo validators are thorough.  

A few things to consider:

1. A load phase and an invoke phase on the worker would let you just load and cache the plugin once and then use it in the invoke.  The initialization of the plugin on every call could be costly and limit opportunities to store state.
2. It may be worth trying protobufs over json just to see if there are speed benefits.
3.  There are a few security concerns around loading the class objects and passing in the system environment to the process.

See more below.

Awesome job!

cpex/framework/models.py
tedhabeck: Unable to get past the ruff check w/o it.
```
   make lint-fix
🔧 Fixing lint issues in cpex...
⚡ Fixing ruff issues in cpex...
error[F821]: Undefined name `List`
    --> cpex/framework/models.py:1840:15
     |
1838 |     latest: Optional[PluginVersionInfo] = None
1839 |     latest_prerelease: Optional[PluginVersionInfo] = None
1840 |     versions: List[PluginVersionInfo]
     |               ^^^^
1841 |
1842 |     def get_version(self) -> Optional[PluginVersionInfo]:
     |
```
leaving it in.

cpex/framework/isolated/worker.py
tedhabeck: The worker now caches the plugin if the config and module_path are unchanged.

cpex/framework/isolated/venv_comm.py
tedhabeck: I added an env parameter which only supplies the PLUGINS_CONFIG_FILE value.

araujof (changes requested):
Nice work, @tedhabeck ! 

Below are a few things I think we should discuss and change before merging: 

### Security

- [ ] **Arbitrary code loading via `sys.path` + `importlib`**
The worker appends user-controlled `plugin_dirs` to `sys.path` then dynamically imports `class_name` via `importlib.import_module`. If `plugin_dirs` or `class_name` originate from untrusted config, this is a remote code execution vector. 

Recommendation: Validate `plugin_dirs` entries against an allowlist (e.g., must be under a known project root). Validate `class_name` against a pattern or registry.

- [ ] **`subprocess.check_call` with user-controlled requirements file**
```python
subprocess.check_call([self.python_executable, "-m", "pip", "install", "-r", requirements_file])
```
The `requirements_file` path comes from plugin config. A malicious requirements file can install arbitrary packages (supply chain attack) or reference local paths. `pip install` can execute `setup.py` which runs arbitrary code.

Recommendation: Validate requirements file path is within the plugin directory. 

- [ ] **No input size limits on stdio communication**
Both the reader thread and worker `main()` call `readline()` without any size limit. A malicious or buggy worker could send an arbitrarily large line, causing OOM.

Recommendation: Add a maximum line length check before parsing.

- [ ] **Worker process inherits CWD**
The worker runs in `os.getcwd()`. If the host CWD contains sensitive files, the plugin code in the worker has filesystem access to them.

- [ ] **No authentication on the stdio channel**
Any process that can write to the worker's stdin can send commands. This is inherent to the subprocess model but worth noting if the threat model evolves.

### Design

- [ ] **`IsolatedVenvPlugin.__init__` reads config file**
The constructor calls `ConfigLoader.load_config()` to discover `plugin_dirs`. This couples plugin instantiation to filesystem config and makes unit testing harder. The plugin dirs should be passed in via `PluginConfig` or injected.

- [ ] **Hardcoded `PLUGINS_CONFIG_FILE` env var fallback**
Both `client.py` and `worker.py` default to `"plugins/config.yaml"`. This is duplicated and fragile. Use PluginSettings to define access and defaults to environment variables.

- [ ] **`to_json()` method on `PluginConfig` manually excludes validator names**
```python
methods_to_exclude = {"_migrate_legacy_modes", "check_url_or_script_filled", ...}
```
Adding a new validator requires updating this set. `model_dump(mode="json")` already excludes methods; the filter is unnecessary since Pydantic v2 doesn't serialize validators.

### Bugs

- [ ] **`sys.path` pollution in worker**
Every call to `process_task` appends to `sys.path` without checking for duplicates. Over many invocations this grows unboundedly.
```python
for module_path in module_paths:
    path = Path(module_path).resolve()
    sys.path.append(resolved_module_path)  # appended every time
```

- [ ] **`module_path` variable leak in `TaskProcessor.initialize`**
```python
tp.initialize(..., module_path=module_path)
```
`module_path` here is the loop variable from the `for module_path in module_paths` loop, so it's always the *last* directory. This may not be the intended one.

- [ ] **Error response in worker uses stale `task_data`**
```python
except json.JSONDecodeError as e:
    error_response = {
        "request_id": task_data.get("request_id", "unknown") if "task_data" in locals() else "unknown",
    }
```
If JSON parsing fails, `task_data` is from the *previous* iteration (or undefined). The `"task_data" in locals()` check is unreliable since the variable persists across loop iterations.

### Nits

- [ ] `print()` statements in `create_venv` should be `logger.info()`.

- [ ] Mixed `orjson` and `json` in `venv_comm.py` -- pick one.

- [ ] `get_environment_info()` in worker.py uses deprecated `importlib.metadata.entry_points()` pattern and slices to `[:10]` arbitrarily.

## Tests

- [ ] No tests for the actual subprocess lifecycle (start/send/stop without mocks)
- [ ] No tests for concurrent `send_task` calls (the request-id routing is untested under concurrency)
- [ ] No negative tests for `sys.path` pollution or the `module_path` variable leak
- [ ] `PluginPackageInfo` and `PluginVersionRegistry` models have no dedicated tests in this PR

araujof (approved):
Nice work addressing our reviews, @tedhabeck !

### What's been resolved                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                            
Addressed security and bug items: 
- arbitrary code loading 
- requirements file traversal
- unbounded stdio reads
- sys.path pollution 
- module_path variable leak
- stale task_data reference 

The to_json() design smell was also cleaned up, and the `PluginPackageInfo`/`PluginVersionRegistry` models now have dedicated test coverage.              
                                                    
###  Acceptable to defer
                                                    
- Hardcoded plugins/config.yaml fallback: The worker dependency was removed, but venv_comm.py:492 still passes os.environ.get("PLUGINS_CONFIG_FILE", "plugins/config.yaml") to the subprocess. This should use PluginSettings for the default rather than a      
  hardcoded string.
- Deprecated entry_points() pattern: get_environment_info() still uses the deprecated calling convention and an arbitrary [:10] slice. Minor but easy to fix.                                                                                                       
- Worker inherits CWD: Low risk given the allowlist validation now in place, but worth a follow-up if the threat model tightens.                                                                                                                                     
- Concurrency and negative tests: The request-id routing and sys.path dedup logic are now correct, but remain undertested under concurrent load. A follow-up issue for integration/stress tests would be appropriate.

PR #8, opened 2026-03-20
feat: capabilities and common message format

### LINKED ISSUE #4: [FEATURE]: Introduce Common Message Format for plugins
### Description

Implement the initial revision of the a Common Message Format (CMF) that defines a **provider-agnostic, structured representation** for interactions between users, agents, tools, and language models.

CMF provides a canonical message model that supports interoperability, policy enforcement, access control, data governance, and end-to-end auditing across heterogeneous model providers and agent frameworks.

[CMF Specification](https://github.com/contextforge-org/contextforge-plugins-framework/blob/652ea1d7311460c09d9502ce0446702bfa53d10c/docs/specs/cmf-message-spec.md)

terylt:
### Summary

As the plugin framework grows to support multi-tenant deployments, security-sensitive workloads, and diverse authentication methods, plugins increasingly need access to contextual metadata: who is making the request, what security labels apply, which HTTP headers were sent, and what agent is orchestrating the call. Today, this metadata is either unavailable to plugins or scattered across payload fields (such as the now-deprecated headers). Extensions address this by providing a single, structured envelope of contextual metadata that travels alongside every hook invocation. But not every plugin should see everything: a logging plugin does not need the user's security clearance, and no plugin should be able to forge identity claims. Capabilities govern this: each plugin declares what it can read and write, and the framework enforces those boundaries automatically. Together, extensions and capabilities give plugin authors rich context while giving platform operators fine-grained control over what each plugin can access and modify.

[CMF Specification](https://github.com/contextforge-org/contextforge-plugins-framework/blob/652ea1d7311460c09d9502ce0446702bfa53d10c/docs/specs/cmf-message-spec.md)

Closes: #4

### Changes

- Introduces a typed Extensions system (`cpex/framework/extensions/`) with optional metadata slots: security (subject, labels, classification, objects, data), HTTP headers, agent context, MCP metadata, provenance, request context, LLM parameters, and custom fields. These flow through the plugin pipeline as structured, frozen Pydantic models.
- Implements a capability-based access control tier system with three mutability levels:
  - **Immutable** (request, provenance, subject, objects, data, classification): plugins can read with the right capability but never modify.
  - **Monotonic** (security labels): plugins can append but never remove, enforced via superset validation on merge.
  - **Mutable** (HTTP headers, agent, custom): plugins with write capabilities can freely modify.
- Build-up filtering (`filter_extensions`) constructs a new `Extensions` object from empty, copying in only the slots granted by the plugin's declared capabilities. Secure by default: new slots remain hidden unless explicitly included.
- Selective merge (`merge_extensions`) controls what the manager accepts back from plugins: immutable slots are silently ignored, monotonic slots are validated for superset, and mutable slots are accepted. No costly before/after diff is needed.
- `SlotName` enum and `FIELD_*` constants in `constants.py` provide a single source of truth for all slot registry keys and Pydantic field names, preventing typo-induced duplicates.
- Optional third-argument dispatch: plugins opt into receiving extensions by adding an `extensions` parameter to their hook method. The signature is inspected once at registration time and stored as a boolean on `HookRef`, with a single branch at dispatch. All 30+ existing 2-parameter plugins continue to work unchanged with zero overhead. Extensions are filtered per plugin via `filter_extensions(extensions, plugin.capabilities)` at the dispatch site.
- Extensions are passed separately from the payload. The payload stays immutable and shared across all plugins (zero copies unless a plugin modifies it); only the small `Extensions` object is filtered/copied per plugin. This maps directly to Rust's `Arc<Payload>` + per-plugin stack-value pattern.
- Adds the Common Message Format (CMF): frozen `Message` and `MessageView` models for multi-modal conversations, with message hooks (`message_pre_invoke`, `message_post_invoke`) for pipeline processing.
- Adds deprecation warnings on redundant header fields across HTTP, agent, and tool payloads (superseded by `extensions.http.headers`), using `warnings.warn(DeprecationWarning)`, which fires once per call site.

### Test Plan

- `python -m pytest tests/ -x -q --tb=short` — all 1159 tests pass
- Verify existing 2-parameter plugins work unchanged (no extensions argument required)
- Verify 3-parameter plugins receive filtered extensions based on declared capabilities
- Verify tier enforcement: immutable slots ignored on merge, monotonic validated for superset, mutable accepted
- Verify `filter_extensions` builds up from empty; undeclared capabilities produce `None` slots
- Verify `merge_extensions` silently drops immutable slot changes and rejects label removal
- Verify deprecation warnings fire once per call site (no log spam)

### DIFF
--- a/cpex/framework/base.py
+++ b/cpex/framework/base.py
@@ -455,7 +455,7 @@ class HookRef:
 
                 # Check for @hook decorator metadata
                 metadata = get_hook_metadata(method)
-                if metadata and metadata.hook_type == hook:
+                if metadata and metadata.matches(hook):
                     self._func = method
                     break
 
--- a/cpex/framework/base.py
+++ b/cpex/framework/base.py
@@ -65,6 +65,11 @@ class Plugin(ABC):
     ) -> None:
         """Initialize a plugin with a configuration and context.
 
+        The plugin receives the config directly. When the plugin is
+        registered with the Manager, the PluginRef retains the
+        authoritative config and gives the plugin a defensive copy,
+        so the Manager never trusts config read back from the plugin.
+
         Args:
             config: The plugin configuration
             hook_payloads: optional mapping of hookpoints to payloads for the plugin.
@@ -267,11 +272,18 @@ class PluginRef:
         ['ref', 'test']
     """
 
-    def __init__(self, plugin: Plugin):
+    def __init__(self, plugin: Plugin, trusted_config: PluginConfig | None = None):
         """Initialize a plugin reference.
+
+        Stores the authoritative config separately from the plugin.
+        The Manager reads policy-sensitive fields (capabilities, mode,
+        on_error) from the trusted config, never from the plugin.
 
         Args:
             plugin: The plugin to reference.
+            trusted_config: The authoritative config retained by the
+                Manager. If not provided, falls back to plugin.config
+                (for backward compatibility in tests).
 
         Examples:
             >>> from cpex.framework import PluginConfig
@@ -293,6 +305,7 @@ class PluginRef:
             True
         """
         self._plugin = plugin
+        self._trusted_config = trusted_config or plugin.config
         self._uuid = uuid.uuid4()
 
     @property
@@ -303,6 +316,15 @@ class PluginRef:
             The underlying plugin.
         """
         return self._plugin
+
+    @property
+    def trusted_config(self) -> PluginConfig:
+        """Return the authoritative config held by the Manager.
+
+        Returns:
+            The trusted PluginConfig (not the plugin's copy).
+        """
+        return self._trusted_config
 
     @property
     def uuid(self) -> str:
@@ -320,7 +342,7 @@ class PluginRef:
         Returns:
             Plugin's priority.
         """
-        return self._plugin.priority
+        return self._trusted_config.priority
 
     @property
     def name(self) -> str:
@@ -329,7 +351,7 @@ class PluginRef:
         Returns:
             Plugin's name.
         """
-        return self._plugin.name
+        return self._trusted_config.name
 
     @property
     def hooks(self) -> list[str]:
@@ -338,7 +360,7 @@ class PluginRef:
         Returns:
             Plugin's configured hooks.
         """
-        return self._plugin.hooks
+        return self._trusted_config.hooks
 
     @property
     def tags(self) -> list[str]:
@@ -347,7 +369,7 @@ class PluginRef:
         Returns:
             Plugin's tags.
         """
-        return self._plugin.tags
+        return self._trusted_config.tags
 
     @property
     def conditions(self) -> list[PluginCondition] | None:
@@ -356,7 +378,7 @@ class PluginRef:
         Returns:
             Plugin's conditions for operation.
         """
-        return self._plugin.conditions
+        return self._trusted_config.conditions
 
     @property
     def mode(self) -> PluginMode:
@@ -365,7 +387,7 @@ class PluginRef:
         Returns:
             Plugin's mode.
         """
-        return self.plugin.mode
+        return self._trusted_config.mode
 
     @property
     def on_error(self) -> OnError:
@@ -374,7 +396,16 @@ class PluginRef:
         Returns:
             Plugin's on_error behavior.
         """
-        return self.plugin.config.on_error
+        return self._trusted_config.on_error
+
+    @property
+    def capabilities(self) -> frozenset[str]:
+        """Return the plugin's declared capabilities.
+
+        Returns:
+            The authoritative capability set from the trusted config.
+        """
+        return self._trusted_config.capabilities
 
 
 class HookRef:
@@ -439,19 +470,25 @@ class HookRef:
             )
 
         # Validate hook method signature (parameter count and async)
-        self._validate_hook_signature(hook, self._func, plugin_ref.plugin.name)
+        param_count = self._validate_hook_signature(hook, self._func, plugin_ref.plugin.name)
 
-    def _validate_hook_signature(self, hook: str, func: Callable, plugin_name: str) -> None:
+        # Store whether the plugin accepts extensions as a third argument
+        self._accepts_extensions = param_count == 3
+
+    def _validate_hook_signature(self, hook: str, func: Callable, plugin_name: str) -> int:
         """Validate that the hook method has the correct signature.
 
         Checks:
-        1. Method accepts correct number of parameters (self, payload, context)
+        1. Method accepts 2 parameters (payload, context) or 3 (payload, context, extensions)
         2. Method is async (returns coroutine)
 
         Args:
             hook: The hook type being validated
             func: The hook method to validate
             plugin_name: Name of the plugin (for error messages)
+
+        Returns:
+            The number of parameters (2 or 3).
 
         Raises:
             PluginError: If the signature is invalid
@@ -462,14 +499,16 @@ class HookRef:
         sig = inspect.signature(func)
         params = list(sig.parameters.values())
 
-        # Check parameter count (should be: payload, context)
+        # Check parameter count (should be: payload, context[, extensions])
         # Note: 'self' is not included in bound method signatures
-        if len(params) != 2:
+        if len(params) not in (2, 3):
             raise PluginError(
                 error=PluginErrorModel(
                     message=f"Plugin '{plugin_name}' hook '{hook}' has invalid signature. "
-                    f"Expected 2 parameters (payload, context), got {len(params)}: {list(sig.parameters.keys())}. "
-                    f"Correct signature: async def {hook}(self, payload: PayloadType, context: PluginContext) -> ResultType",
+                    f"Expected 2 or 3 parameters (payload, context[, extensions]), "
+                    f"got {len(params)}: {list(sig.parameters.keys())}. "
+                    f"Correct signature: async def {hook}(self, payload: PayloadType, "
+                    f"context: PluginContext[, extensions: Extensions]) -> ResultType",
                     plugin_name=plugin_name,
                 )
             )
@@ -484,6 +523,8 @@ class HookRef:
                     plugin_name=plugin_name,
                 )
             )
+
+        return len(params)
 
         # ========== OPTIONAL: Type Hint Validation ==========
         # Uncomment to enable strict type checking of payload and return types.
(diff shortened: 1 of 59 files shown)


araujof (changes requested):
Nice work, @terylt!

This PR introduces two major subsystems with clean separation of concerns:

- Extensions (capability-gated metadata)
- CMF (Common Message Format)

The design looks good:

- **Extension filtering** (`filter_extensions`) is secure-by-default: new slots are hidden unless explicitly registered.
- **Mutability controls** (immutable / monotonic / mutable) with enforcement at both filter and merge boundaries.
- **Trusted config separation** via `PluginRef._trusted_config` prevents plugins from escalating their own capabilities post-registration. `PluginConfig` is now frozen.
- **Optional 3rd-argument to invoke method** preserves backward compatibility.

Below are a findings that we should discuss and address before merging:

### Security

**1. `DelegationPayload.bearer_token` stored in plaintext**

`DelegationPayload.bearer_token` (`hooks/identity.py`) carries the caller's current bearer token as a plain `str` field in a Pydantic model. This token will be serialized to JSON if the payload is logged, audited, or serialized for external plugins (gRPC/MCP). Consider:
- A `SecretStr` type or a custom field that redacts on serialization
- Ensuring fire-and-forget and audit-mode plugins don't receive this field

**2.`IdentityPayload.raw_token` same concern**

`IdentityPayload.raw_token` carries the raw credential (JWT, API key, etc.). Same serialization risk as 1.

**3. `datetime.utcnow()` deprecated**

`delegation.py` use `datetime.utcnow()` which is deprecated since Python 3.12 and returns a naive datetime. Use `datetime.now(datetime.UTC)` instead.

**4. `to_dict()` sensitive header stripping is case-sensitive at the check boundary**

`view.py` line `_SENSITIVE_HEADERS = frozenset({"authorization", "cookie", "x-api-key"})` lowercases the check, but `HttpExtension.headers` is `dict[str, str]` with no case normalization. Headers like `Authorization` are correctly caught by the `k.lower()` comparison in `to_dict()`, but the `headers` property on `MessageView` returns them as-is. This is fine for `to_dict()` but plugins accessing `view.headers` directly get the raw sensitive values. This is by design (capability-gated), but worth documenting explicitly.

**5. `ConversationContext.history` typed as `list[Any]`**

`agent.py`, `ConversationContext.history: list[Any]` is an untyped escape hatch. Arbitrary objects in this field could carry serialization gadgets or unexpectedly large payloads. Consider constraining the type or adding size limits.

### Code Quality

**6. Dead code after `return` in `_validate_hook_signature`**

`base.py:524` — the `return len(params)` statement is followed by commented-out type hint validation code. The `return` makes everything after it unreachable. The commented block should either be removed or placed before the return.

**7.  `_serial_phase_state` removal is clean but return tuple is complex**

The removal of `self._serial_phase_state` mutable state in favor of returning a tuple `(halt_result, (payload, decision_plugin_name, extensions))` is a good improvement. However, the nested tuple return type is complex. A small `@dataclass` (e.g., `PhaseState`) would improve readability.

**8. `merge_extensions` doesn't merge `delegation` slot**

`tiers.py` `merge_extensions()` handles `http`, `security`, and `custom` but does not handle the `delegation` slot at all. `DelegationExtension` has monotonic semantics (chain grows), but it's not in the slot registry and not handled in merge. If a plugin modifies delegation through the extensions object, changes will be silently dropped.

**9. `HttpHeaderPayload` no longer inherits from `PluginPayload`**

`http.py` changed `HttpHeaderPayload(RootModel[dict[str, str]], PluginPayload)` to `HttpHeaderPayload(RootModel[dict[str, str]])`. This is a breaking change for any code that checks `isinstance(payload, PluginPayload)` on this type. Was that intentional?

**10. Deprecation warnings fire on every instantiation with headers**

The `@field_validator("headers", mode="before")` deprecation warnings in `HttpPreRequestPayload`, `ToolPreInvokePayload`, `AgentPreInvokePayload`, and `HttpAuthResolveUserPayload` fire whenever `headers` is not `None`. Since `headers` is a commonly-used required field in HTTP payloads, this will be noisy for existing users who haven't migrated to extensions yet. The `stacklevel=2` may not point to the right caller since Pydantic validators are called from within Pydantic's validation machinery. Consider using `warnings.warn(..., stacklevel=4)` or testing the actual warning output.

**11. Multi-hook decorator (`@hook([...])`) — HookRef matches first**

`HookRef.__init__` uses `metadata.matches(hook)` which checks `hook_type in self.hook_types`. When a method is decorated with `@hook([CmfHookType.TOOL_PRE_INVOKE, CmfHookType.TOOL_POST_INVOKE])`, the same method object is reused for both hooks. This is fine functionally but means both HookRefs share the same underlying bound method. Confirm that the method can correctly handle both payload shapes if they differ.

**12. `Message.extensions` field on the model vs separate passing**

`Message` has no `extensions` field in its model definition (extensions are passed separately to `iter_views()`). However, `test_view.py` constructs `Message(role=..., content=..., extensions=ext)`. Since `Message` uses `ConfigDict(frozen=True)` without `extra="forbid"`, this silently passes (Pydantic ignores extra fields by default with frozen).

### Missing Test Coverage

- `delegation.py`: `DelegationExtension.with_new_hop()` has no dedicated tests
- `identity.py`: `IdentityPayload`, `DelegationPayload`, `AttenuationConfig` have no tests
- No integration test verifying end-to-end flow: plugin receives filtered extensions, modifies allowed slots, manager merges back
- No test for `_execute_with_timeout` with `accepts_extensions=True` path through the manager

terylt:
Thanks for the thorough review — all items addressed. Here's the summary:                                                        
                                                                                                                                   
  Security (1-3, 8)                     
                                                                                                                                   
  - IdentityPayload.raw_token and DelegationPayload.bearer_token now use SecretStr — redacted on serialization, .get_secret_value()
   for actual access                                                                                                               
  - datetime.utcnow() replaced with datetime.now(UTC) — timestamps are timezone-aware                                              
  - Delegation added to slot registry as monotonic + capability-gated (read_delegation / append_delegation). merge_extensions()    
  validates chain growth (no shrink, no prefix tampering). filter_extensions() hides delegation from plugins without               
  read_delegation                                                                                                                  
                                                                                                                                   
  Code Quality (6, 7, 9, 12)                                                                                                   
                                                                                                                                   
  - Dead code after return in _validate_hook_signature removed                                                                     
  - PhaseState dataclass replaces the nested tuple[Optional[PluginPayload], Optional[str], Optional[Extensions]] return from
  _run_serial_phase                           
  - HttpHeaderPayload restored to inherit from PluginPayload — tested and confirmed compatible with RootModel. The original removal
   was unnecessary                            
  - Message now has extra="forbid" — extra fields raise ValidationError. Fixed one test that was silently passing extensions to    
  Message() instead of iter_views()           
                                                                                                                                   
  Documentation/Minor (4, 5, 10)
                                                                                                                                   
  - view.headers property now documents that sensitive headers are returned raw (by design for capability-gated plugins) and only
  stripped in to_dict()                                                                                                            
  - ConversationContext.history constrained from list[Any] to list[BaseModel] with max_length=100
  - Deprecation warnings stacklevel changed from 2 to 4 to point past Pydantic's validation machinery                              
                                              
  Intentional — No Change (11)                                                                                                    
                                                                                                                                   
  - Multi-hook shared method confirmed correct — all CMF hooks use the same MessagePayload type                                    
                                                                                                                                   
  Test Coverage Added (48 new tests, 1502 total)                                                                                   
                                                                                                                                   
  - test_delegation.py (11) — DelegationHop construction, DelegationExtension.with_new_hop() monotonic chain growth, immutability,
  age calculation                                                                                                                  
  - test_identity.py (14) — IdentityPayload/DelegationPayload SecretStr redaction, AttenuationConfig, IdentityResult
  - test_extensions_integration.py (16) — end-to-end filter → modify → merge for labels, delegation, custom, headers with/without  
  capabilities                                
  - test_manager_extensions.py (6) — _execute_with_timeout with accepts_extensions=True, plugin receives filtered extensions,      
  plugin modifies labels/custom and manager merges back                                                                            
  - Updated test_tiers.py for new READ_DELEGATION/APPEND_DELEGATION capabilities

araujof (approved):
LGTM! Nice work!

## Verified fixes

| # | Concern | Status | Evidence |
|---|---|---|---|
| 1 | `IdentityPayload.raw_token` stored in plaintext | Fixed | `SecretStr` at `cpex/framework/hooks/identity.py:90` |
| 2 | `DelegationPayload.bearer_token` stored in plaintext | Fixed | `SecretStr` at `cpex/framework/hooks/identity.py:219` |
| 3 | `datetime.utcnow()` deprecated | Fixed | `datetime.now(UTC)` at `cpex/framework/extensions/delegation.py:56,113` |
| 4 | `view.headers` returns raw sensitive values | Documented | Explicit `.. note::` at `cpex/framework/cmf/view.py:732–747` |
| 5 | `ConversationContext.history: list[Any]` escape hatch | Fixed | Constrained to `list[BaseModel]` with `max_length=100` at `cpex/framework/extensions/agent.py:44` |
| 6 | Dead code after `return` in `_validate_hook_signature` | Fixed | Commented block extracted into proper `_validate_type_hints` method in `cpex/framework/base.py` |
| 7 | Complex nested tuple return from `_run_serial_phase` | Fixed | `@dataclass PhaseState` at `cpex/framework/manager.py:79` |
| 8 | `merge_extensions` doesn't handle `delegation` slot | Fixed | `SlotName.DELEGATION` registered with `READ_DELEGATION`/`APPEND_DELEGATION` capabilities; superset + prefix-immutability validation at `cpex/framework/extensions/tiers.py:613–644` |
| 9 | `HttpHeaderPayload` no longer inherits `PluginPayload` | Fixed | Restored to `RootModel[dict[str, str]], PluginPayload` at `cpex/framework/hooks/http.py:21` |
| 10 | Deprecation warnings `stacklevel=2` wrong frame | Fixed | `stacklevel=4` across `cpex/framework/hooks/http.py`, `hooks/tools.py`, `hooks/agents.py` |
| 11 | Multi-hook decorator shares bound method | Declared intentional | All CMF hooks use the same `MessagePayload` type — no divergent payload shapes |
| 12 | `Message` silently accepts extra fields | Fixed | `extra="forbid"` at `cpex/framework/cmf/message.py:893`; test bug (extensions passed to `Message()` instead of `iter_views()`) corrected |

## Test coverage additions

| Test file | Tests | Covers |
|---|---|---|
| `tests/unit/cpex/framework/extensions/test_delegation.py` | 11 | `DelegationHop` construction, `with_new_hop()` monotonic chain growth, immutability, age calculation |
| `tests/unit/cpex/framework/hooks/test_identity.py` | 14 | `IdentityPayload`/`DelegationPayload` `SecretStr` redaction, `AttenuationConfig`, `IdentityResult` |
| `tests/unit/cpex/framework/test_extensions_integration.py` | 17 | End-to-end filter → modify → merge for labels, delegation, custom, headers (with/without capabilities) |
| `tests/unit/cpex/framework/test_manager_extensions.py` | 6 | `_execute_with_timeout` with `accepts_extensions=True`, plugin receives filtered extensions, manager merges back |
| **Total new tests** | **48** | — |

## Conclusion

All actionable review items are resolved in code; item 11 was deemed intentional with a documented rationale. The security-critical fixes (credential redaction via `SecretStr`, delegation slot registration with monotonic chain enforcement, superset validation on merge) are implemented. The added tests close the gaps called out in the review, including the previously-missing end-to-end integration path through the manager with `accepts_extensions=True`.

PR #33, opened 2026-04-21
feat: expose background_tasks in PluginResult for fire-and-forget synchronization

### LINKED ISSUE #25: [FEATURE]: Expose FIRE_AND_FORGET task handles on PluginResult for deterministic awaiting
### Summary

`PluginExecutor._fire_and_forget_tasks()` schedules background work via
`asyncio.create_task()` but discards the resulting task handles. Callers of
`PluginManager.invoke_hook()` have no way to know when those background tasks
complete. The only workaround today is an arbitrary `asyncio.sleep()`, which
is racy under load or on slow CI runners.

### User Story 1

* **As a:** developer using CPEX's `FIRE_AND_FORGET` plugin mode
* **I want:** the task handles created by `_fire_and_forget_tasks()` to be accessible on `PluginResult`
* **So that:** I can deterministically await background task completion in tests and other sync-sensitive contexts

#### Acceptance Criteria

```gherkin
Scenario: Caller awaits background tasks after hook invocation
  Given a hook with one or more FIRE_AND_FORGET plugins registered
  When invoke_hook() returns a PluginResult
  Then result.background_tasks contains the asyncio.Task handles created for those plugins

Scenario: Callers that ignore background_tasks see no behavior change
  Given a hook invocation with no caller reading background_tasks
  When invoke_hook() completes
  Then fire-and-forget tasks still run in the background as before
```

### Design Sketch (optional)

Add a `background_tasks` field to `PluginResult`:

```python
class PluginResult(BaseModel, Generic[T]):
    ...
    background_tasks: list[asyncio.Task[Any]] = Field(default_factory=list)
```

Populate it in `_fire_and_forget_tasks()` by collecting the return values of
`asyncio.create_task()`. Callers that need a deterministic sync point can then:

```python
result, payload = await manager.invoke_hook(...)
await asyncio.gather(*result.background_tasks, return_exceptions=True)
```

### Alternatives Considered (optional)

- **Fixed `asyncio.sleep()`** — current workaround; racy under load, arbitrary timeout.
- **`asyncio.all_tasks()` snapshot** — capture tasks before/after `invoke_hook()`; fragile (captures unrelated tasks, depends on internal implementation detail).
- **Test-only `asyncio.create_task` patch** — monkey-patching asyncio in tests; messy and global.

### Additional Context

Surfaced while writing deterministic tests for telemetry plugins in a downstream
project that uses CPEX's `FIRE_AND_FORGET` mode. See:
https://github.com/generative-computing/mellea/issues/691

ajbozarth:
### Summary

Adds a `background_tasks` field to `PluginResult` containing the `asyncio.Task`
handles created for `FIRE_AND_FORGET` plugins. Callers can now await background
tasks deterministically instead of using arbitrary sleep delays.

Closes: #25

### Changes

- `PluginResult` gains a `background_tasks: list[asyncio.Task]` field (excluded
  from serialization; defaults to empty list for backward compatibility)
- `_fire_and_forget_tasks()` now returns the list of task handles instead of `None`
- Both the normal execution path and the halt path in `execute()` populate
  `result.background_tasks`
- Five tests in `test_plugin_modes.py` updated to use
  `await asyncio.gather(*result.background_tasks, return_exceptions=True)`
  in place of `asyncio.sleep()` workarounds
- `PluginResult` entry in `docs/specs/plugin-framework-spec.md` updated

### Checks

- [x] `make lint` passes
- [x] `make test` passes

### Notes

`arbitrary_types_allowed=True` is added to `PluginResult.model_config` to
allow Pydantic to accept `asyncio.Task` as a field type — the same config
already used on `PluginPayload`. The `exclude=True` on the field keeps
`.model_dump()` / `.model_dump_json()` output unchanged.

### DIFF
--- a/cpex/framework/manager.py
+++ b/cpex/framework/manager.py
@@ -354,7 +354,7 @@ class PluginExecutor:
                     )
 
         # FIRE_AND_FORGET: fire-and-forget background tasks (fires last with final payload snapshot)
-        self._fire_and_forget_tasks(
+        bg_tasks = self._fire_and_forget_tasks(
             fire_and_forget_refs,
             payload,
             global_context,
@@ -373,6 +373,7 @@ class PluginExecutor:
                 modified_extensions=current_extensions,
                 violation=None,
                 metadata=combined_metadata,
+                background_tasks=bg_tasks,
             ),
             res_local_contexts,
         )
@@ -688,7 +689,7 @@ class PluginExecutor:
         extensions: Optional[Extensions] = None,
     ) -> tuple[PluginResult, dict]:
         """Schedule fire-and-forget tasks and build a pipeline-halting result."""
-        self._fire_and_forget_tasks(
+        bg_tasks = self._fire_and_forget_tasks(
             fire_and_forget_refs,
             payload,
             global_context,
@@ -704,6 +705,7 @@ class PluginExecutor:
                 modified_payload=current_payload,
                 violation=violation,
                 metadata=combined_metadata,
+                background_tasks=bg_tasks,
             ),
             res_local_contexts,
         )
@@ -728,12 +730,14 @@ class PluginExecutor:
         res_local_contexts: dict,
         semaphore: Optional[asyncio.Semaphore],
         extensions: Optional[Extensions] = None,
-    ) -> None:
+    ) -> list[asyncio.Task]:
         """Schedule all FIRE_AND_FORGET plugins as fire-and-forget background tasks.
 
         May be called from an early-exit path or from the normal completion path.
         Each FIRE_AND_FORGET plugin receives an isolated snapshot of the payload at call time.
+        Returns the list of asyncio.Task handles for all newly scheduled tasks.
         """
+        tasks: list[asyncio.Task] = []
         for ref in fire_and_forget_refs:
             local_context_key = global_context.request_id + ref.plugin_ref.uuid
             if local_context_key in res_local_contexts:
@@ -752,9 +756,11 @@ class PluginExecutor:
             )
             local_context = PluginContext(global_context=tmp_gc)
             res_local_contexts[local_context_key] = local_context
-            asyncio.create_task(
+            task = asyncio.create_task(
                 self._run_fire_and_forget_task(ref, task_input, local_context, semaphore, extensions=extensions)
             )
+            tasks.append(task)
+        return tasks
 
     async def _run_fire_and_forget_task(
         self,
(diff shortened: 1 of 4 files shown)


araujof (approved):
Thanks for this PR @ajbozarth! Nice work. This is a good change that preserves backwards compatibility. The test migration from `asyncio.sleep()` to direct task awaits is also a good win for determinism and CI reliability.

### What I verified

- All scheduling paths (`execute()` normal completion + `_build_halt_result` halt path) correctly populate `background_tasks`.
- `exclude=True` keeps the field out of `model_dump()` / `model_dump_json()`, so MCP, gRPC, and isolated-worker serialization are unaffected.
- `arbitrary_types_allowed=True` is consistent with the existing pattern on `PluginPayload`.
- Task lifecycle is improved, callers now hold strong references to fire-and-forget tasks via the result.

### Worth noting

- Remote callers (MCP/gRPC/isolated) always receive `background_tasks=[]`; they can't synchronize with tasks in the remote process. Not a bug, but worth a docstring mention if remote sync ever becomes a requirement.
- Callers now have `.cancel()` access on fire-and-forget tasks. If that's undesirable, wrapping in `asyncio.shield` or returning `Sequence[Awaitable]` would restrict it. Not required today.

@terylt can you review to see if you spot any issues, specially any undesirable interactions with the upcoming Rust Core for CPEX?

terylt:
Overall @ajbozarth Nice work and thanks for the contribution!

I have one suggestion just to help keep parity with the rust core we are building out. Rather than exposing `background_tasks` as the primary API, consider adding a   `wait_for_background_tasks()` method on `PluginResult` that wraps the `asyncio.gather` call and returns  `list[PluginError]` — an empty list on success, populated with errors from any tasks that failed. Callers would use   `errors = await result.wait_for_background_tasks()` instead of manually gathering the task list. 

The internal list can   stay for advanced use cases, but the method gives a cleaner public API and a consistent return type for error  handling. This also aligns with the Rust core — `PipelineResult` will expose the same `wait_for_background_tasks()`   method backed by `tokio::JoinHandle`s, converting failures to `PluginError` before returning. The PyO3 wrapper will   map it to a Python awaitable with the same signature. Keeping the API shape consistent across both runtimes means  plugins and tests work the same way regardless of which backend is running.

ajbozarth:
Thanks for the feedback — addressed in the second commit.

`_run_fire_and_forget_task` now returns `PluginErrorModel | None` instead
of swallowing exceptions, so errors survive the task boundary.
`wait_for_background_tasks()` on `PluginResult` does the gather internally
and returns `list[PluginErrorModel]` — empty on success, one entry per
failed task. All five tests updated to use it, plus a new test that asserts
on the returned error model directly. Spec doc updated to match.

terylt (approved):
LGTM. Nice work @ajbozarth!

PR #38, opened 2026-04-22
feat: CPEX Rust config

### LINKED ISSUE #16: [FEATURE]: YAML config parsing for RUST Core with routing rules.
**Epic**: CPEX Rust Core (#12)

## Summary

Implement the unified YAML config parser in `cpex-core` (`config.rs`). Parses the single config format that combines global policies, defaults, plugin declarations, and per-entity routes into typed Rust structs via serde.

## Motivation

Currently, plugins must be registered programmatically. Config-driven registration enables declarative plugin setup from a single YAML file.  We will add routing rules to the current plugin configuration, and provide experimental integrations with APL eventually.

## Scope

- `CpexConfig` top-level struct: `global`, `plugins`, `routes`
- `GlobalConfig`: `policies` (named policy groups with `all` reserved), `defaults` (per entity type), `delegation`, `identity`, `session`
- `PluginDeclConfig`: per-plugin declaration (name, kind, hooks, mode, priority, capabilities, conditions, config)
- `RouteEntry`: entity matcher (tool/resource/prompt/llm), `meta` (tags, scope, properties), `policy`, `post_policy`, `args`, `result`, `taint`
- Validation at load time: duplicate plugin names, unknown hook names, invalid mode/capability values
- Integration with `PluginManager` — config drives `register_handler` calls

## Acceptance Criteria

- Parse the example YAML from the unified config proposal without errors
- Type validation catches malformed configs at startup
- `PluginManager` can be initialized from a config file path
- Route matching: exact name > name list > glob > meta tag match > when expression > defaults

## References

- [Plugin Framework Spec](https://github.com/contextforge-org/contextforge-plugins-framework/blob/feat/capabilities_cmf/docs/plugin-framework-spec-v2.md)
- [unified config proposal](https://github.com/contextforge-org/contextforge-plugins-framework/blob/feat/apl/docs/unified-config-proposal.md)
- [APL DSL spec](https://github.com/contextforge-org/contextforge-plugins-framework/blob/feat/apl/docs/apl-dsl-spec.md)
- Current stub: `crates/cpex-core/src/config.rs`

terylt:
### Summary                                                                                                          
                                                                                                                        
  Adds config-driven plugin management to the CPEX Rust core. Plugins, policy groups, and routing rules are declared in 
  a single YAML config. The manager resolves routes at dispatch time based on entity metadata in extensions, with cached
   lookups for high-throughput use cases.                                                                               
                                                                                                                        
  Built on top of feat/cpex_rust_core (async invoke, PluginContextTable, OnError::Disable, etc.).                       
                                                                                                                        
  **Config parser** — Parses a unified YAML format with `plugins`, `policy_groups`, and `routes` sections. Routes match 
  entities by exact name, name list, glob, or wildcard with specificity-based resolution (exact 1000 > list 500 > glob  
  300 > wildcard 0). Scope filtering, tag merging (union), and `when` clauses (carried for APL evaluation) are          
  supported. Validation at parse time rejects duplicate plugins, unknown plugin refs, and malformed routes.             
                  
  **Factory system** — `PluginFactory` trait maps `kind` strings to plugin constructors. `from_config(yaml, factories)` 
  builds a fully wired `PluginManager` from YAML. Routes can override plugin config; the factory creates new instances
  with merged config.
                                                                                                                        
  **Route-filtered dispatch** — `invoke_by_name` and `invoke::<H>` resolve routes via `MetaExtension` (entity_type, 
  entity_name, scope, tags) before dispatching. When routing is disabled or meta is absent, all registered plugins fire.
                  
  **Route cache** — `RwLock<HashMap<RouteCacheKey, Arc<Vec<HookEntry>>>>` with hashbrown `raw_entry` for zero-allocation
   cache hits and `Arc` for zero-copy reads. Keyed by (entity_type, entity_name, hook_name, scope). 
  `clear_route_cache()` for config reload.                                                                              
                                                                                                                        
  **MetaExtension** — Added to `Extensions` with entity_type, entity_name, tags, scope, and properties. Used by the 
  routing system to identify which entity a hook invocation targets.     

Closes: #16 

### Changes

- `config.rs` — Unified YAML config parser with plugins, policy groups, routes, specificity-based resolution, scope   
  matching, tag merging, and validation                                                                                 
  - `factory.rs` — `PluginFactory` trait and `PluginFactoryRegistry` for config-driven plugin instantiation             
  - `manager.rs` — `from_config`/`load_config`, route-filtered dispatch, hashbrown route cache with `Arc` zero-copy     
  reads, override instances with merged config, `clear_route_cache`                                                     
  - `hooks/payload.rs` — `MetaExtension` (entity_type, entity_name, tags, scope, properties) on `Extensions`            
  - `registry.rs` — `register_for_names_with_handler` for factory path (no `HookTypeDef` type parameter required)       
  - `lib.rs` — Added `factory` module                                                                                   
  - `Cargo.toml` — Added `hashbrown` dependency

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,12 @@
 version = 4
 
 [[package]]
+name = "allocator-api2"
+version = "0.2.21"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "683d7910e743518b0e34f1186f92494becacb047c7b6bf616c96772180fef923"
+
+[[package]]
 name = "anyhow"
 version = "1.0.102"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -49,6 +55,7 @@ version = "0.1.0"
 dependencies = [
  "async-trait",
  "futures",
+ "hashbrown 0.15.5",
  "serde",
  "serde_json",
  "serde_yaml",
@@ -197,6 +204,8 @@ version = "0.15.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9229cfe53dfd69f0609a49f65461bd93001ea1ef889cd5529dd176593f5338a1"
 dependencies = [
+ "allocator-api2",
+ "equivalent",
  "foldhash",
 ]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,3 +29,4 @@ tracing = "0.1"
 uuid = { version = "1", features = ["v4"] }
 paste = "1"
 futures = "0.3"
+hashbrown = "0.15"
--- a/crates/cpex-core/Cargo.toml
+++ b/crates/cpex-core/Cargo.toml
@@ -25,3 +25,4 @@ thiserror = { workspace = true }
 tracing = { workspace = true }
 uuid = { workspace = true }
 futures = { workspace = true }
+hashbrown = { workspace = true }
--- a/crates/cpex-core/examples/README.md
+++ b/crates/cpex-core/examples/README.md
@@ -1,0 +1,43 @@
+# CPEX Core Examples
+
+## plugin_demo
+
+A complete end-to-end example showing how to build plugins, load config, and invoke hooks with the CPEX runtime.
+
+### What it demonstrates
+
+- **Defining hook types and payloads** — `ToolPreInvoke` and `ToolPostInvoke` hooks with a shared `ToolInvokePayload`
+- **Building plugins** — three plugins (`IdentityResolver`, `PiiGuard`, `AuditLogger`) implementing `Plugin` + `HookHandler<H>` for different hook types
+- **Multi-hook registration** — a single plugin instance (e.g., `IdentityResolver`) registered for multiple hooks (`tool_pre_invoke` and `tool_post_invoke`) via the factory pattern
+- **Plugin factories** — `PluginFactory` implementations that create plugin instances and wire up typed handler adapters
+- **YAML config loading** — `plugin_demo.yaml` declares plugins, policy groups, and routing rules
+- **Policy groups and tag-based routing** — the `pii` policy group activates `PiiGuard` only for tools tagged with `pii`
+- **Route resolution** — exact tool matches, wildcard catch-all, tag-driven plugin selection
+- **PluginContext** — `global_state` used to pass PII clearance between hooks, `local_state` for per-plugin scratch data
+- **BackgroundTasks** — fire-and-forget plugins (`AuditLogger`) spawn background tasks; `wait_for_background_tasks()` awaits them
+- **PluginContextTable** — context table threaded from pre-invoke to post-invoke to preserve plugin state
+
+### Running
+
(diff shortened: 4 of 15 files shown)


araujof (commented):
Nice work, @terylt!

Overall,

- Trust model enforcement is consistent: config loader -> manager -> PluginRef -> executor, never from plugin.
- The `Arc`-based route cache with `raw_entry` zero-allocation hits is a good optimization.
- Test coverage spans all execution modes, error behaviors, routing resolution, and cache behavior.
- The 5-phase executor design is clean and the phase separation is well-enforced.

A few minor comments, let me know if they are planned in a next PR, or if you want to address them here.

### Issues

**1. Glob matching is overly simplistic** (`config.rs`)

`StringOrList::matches()` only handles suffix wildcards (`prefix*`). Was that intentional?

```rust
let prefix = pattern.trim_end_matches('*');
name.starts_with(prefix)
```

This silently mismatches on interior or leading globs. Either document that only trailing `*` is supported, or use a glob crate (e.g., `glob-match`).

**2. `from_config` duplicates `load_config` logic** (`manager.rs`)

`from_config()` and `load_config()` have the same plugin instantiation loop and executor setup, but diverged implementations. Consider having `from_config` delegate to `load_config` internally.

**3. `RwLock::unwrap()` will panic on poisoned lock** (`manager.rs`)

If a thread panics while holding the write lock, all subsequent `unwrap()` calls will panic too, cascading across the system. Since this is a library crate consumed by others, consider using `.read().unwrap_or_else(|e| e.into_inner())` (or `parking_lot::RwLock` which is non-poisoning and already a transitive dep via tokio).

**4. `find_matching_route` returns only the single best route** (`config.rs`)

If two routes have equal specificity for the same entity, the last one in config order wins (due to `total > s` using strict greater-than). This is deterministic but not documented, and could surprise users. Consider either documenting "first wins on tie" or returning all tied routes.

**5. `serde_yaml` is deprecated** (`Cargo.toml`)

The `serde_yaml` crate is deprecated in favor of alternatives. Not blocking, but worth noting for future migration planning (e.g., `serde_yml` or `yaml-rust2` + manual serde).

### Suggestions (non-blocking)

- **`create_override_instance` silently falls back** (`manager.rs`): when factory creation fails, it logs an error and returns `None`, causing the base instance to fire with the original config. This could mask misconfigurations. Consider propagating the error or at minimum emitting a `warn!` with the route context.

- **`FilteredExtensions` is always `default()`** (`executor.rs`): the TODO is fine for Phase 1, but since `meta` is needed for route resolution and plugins might want to read it, you could pass through `meta` in the filtered view now. Currently plugins can't see the meta extension that drives their own dispatch.

- **Shutdown in reverse order** (`manager.rs`): the comment says "reverse registration order" but `plugin_names()` returns insertion order. If you want reverse, call `.rev()` on the iterator.

araujof:
LGTM. Fixes to this PR will fold into #45.

araujof (approved):

PR #42, opened 2026-04-26
fix: handle exception types when isolating payloads

### LINKED ISSUE #41: [BUG]: Unexpected warning when payload fields derive from BaseException
### Current Behavior
CPEX outputs an unnecessary warning when isolating payloads containing `BaseException`-derived fields. Since exceptions are immutable, CPEX should just handle this case without warning. 

### Expected Behavior
No warning should be generated.

### Steps to Reproduce

```python
import logging, warnings
logging.basicConfig(level=logging.DEBUG)
warnings.simplefilter('always')

from openai import APIConnectionError
from cpex.framework.memory import wrap_payload_for_isolation, _wrap_value, _safe_deepcopy

# Build a GenerationErrorPayload with an APIConnectionError
from mellea.plugins.hooks.generation import GenerationErrorPayload

exc = APIConnectionError(request=None)
payload = GenerationErrorPayload(exception=exc)

print('Payload created. exception field type:', type(payload.exception).__qualname__)
print()
print('--- Attempting wrap_payload_for_isolation (what CPEX does for FIRE_AND_FORGET) ---')
isolated = wrap_payload_for_isolation(payload)
print()
print('Isolation completed. Isolated exception is same object?', isolated.exception is exc)
```
### Logs / Error Output
```bash
WARNING  cpex.framework.memory:memory.py:536 Cannot deep-copy value of type APIConnectionError — sharing reference: APIConnectionError.__init__() takes 1 positional argument but 2 were given
```

### Additional Context (optional)
Here is what happens:
1. `GenerationErrorPayload.exception` is typed as BaseException (not a primitive, not a dict, not a list, not a BaseModel)                                                                                                                                                                                                   
2. `wrap_payload_for_isolation` iterates over model fields, hits exception, calls `_wrap_value(exc)`                                                                                                                                                                                                                           
3. `_wrap_value` falls through all type checks to the bottom: return `_safe_deepcopy(value)`                                                                                                                                                                                                                                   
4. `_safe_deepcopy` calls `copy.deepcopy(exc)` which fails because `APIConnectionError.__init__` takes keyword-only args `(*, message, request)` but Python's default `__reduce__` passes `exc.args` as positional args                                                                                                                
5. The warning is logged and the original exception is shared by reference

araujof:
### Summary
CPEX should not try to deep copy exception objects in payload fields.

Closes: #41 

### Changes
One line added to `_wrap_value()` in `cpex/framework/memory.py`: `BaseException` isinstance check that returns the exception object as-is, placed right after the primitives check and before other checks. This prevents `_safe_deepcopy` from ever being called on exception objects.

Why it's safe: Exceptions are effectively immutable in practice. Plugins should not mutate the exception, so sharing by reference is acceptable.

Tests added: 7 new tests in TestExceptionIsolation covering `_wrap_value` with standard exceptions, non-copyable exceptions (keyword-only `__init__`), `BaseException` subclasses, full `wrap_payload_for_isolation` on exception-carrying payloads, no-warning verification, and exception + dict coexistence in the same payload.

### Checks

- [x] `make lint` passes
- [x] `make test` passes

### DIFF
--- a/cpex/framework/memory.py
+++ b/cpex/framework/memory.py
@@ -558,6 +558,8 @@ def _wrap_value(value: Any) -> Any:
         return value
     if isinstance(value, _PRIMITIVE_TYPES):
         return value
+    if isinstance(value, BaseException):
+        return value
     if isinstance(value, RootModel):
         root = value.root
         if isinstance(root, dict):
--- a/tests/unit/cpex/framework/test_memory.py
+++ b/tests/unit/cpex/framework/test_memory.py
@@ -1396,3 +1396,99 @@ class TestSafeDeepCopy:
 
         assert result is lock
         assert "Cannot deep-copy" in caplog.text or "sharing reference" in caplog.text
+
+
+# ---------------------------------------------------------------------------
+# Tests: BaseException isolation (no deepcopy)
+# ---------------------------------------------------------------------------
+
+
+class _NonCopyableError(Exception):
+    """Exception whose __init__ uses keyword-only args, breaking deepcopy."""
+
+    def __init__(self, *, message: str = "error", request: object = None):
+        super().__init__(message)
+        self.request = request
+
+
+class _PayloadWithException(BaseModel):
+    """Payload carrying an exception field (like GenerationErrorPayload)."""
+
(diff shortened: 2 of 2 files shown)


terylt (approved):
LGTM

PR #49, opened 2026-05-07
feat: change Plugin handler to async for performance

terylt:
### Summary                                                                                                            
                                                                                                                        
  Collapses `HookHandler<H>` and `AsyncHookHandler<H>` into a single async-by-default trait. Plugin authors now write   
  `async fn handle(...)` regardless of whether the body actually awaits anything — handlers with no `.await` compile to 
  a trivially-ready future that LLVM inlines, so there's no observable cost over a plain function.                      
                                                                                                                        
 ### Why
                                                                                                                        
  The previous design had `HookHandler<H>` (sync) and `AsyncHookHandler<H>` (async) with a blanket impl auto-promoting
  sync to async. Two traits, one keyword of difference (`fn` vs `async fn`). The only real benefit of keeping them      
  separate was a compile-time forcing function preventing `.await` on the hot path — thin value for the cognitive       
  overhead of two traits in docs, error messages, and bounds.                                                           
  
  One trait, one bound, one signature. Authors decide whether to await based on what they need, not which trait they    
  implement.                              

 ### What changed                                                                                                       
  
  - `HookHandler<H>::handle` is now `async fn` (native AFIT, returns `impl Future + Send`)                              
  - Deleted the sync `HookHandler<H>` trait and the auto-promotion blanket impl
  - All sync plugin impls (test plugins, demos, FFI test plugin, go-demo plugins) gained one `async` keyword
                                       
                                              
### No behavior change                                                                                                 
                                                                                                                        
  - Sync-style plugins (no `.await` in body) compile to the same machine code via inlining                              
  - Async plugins work exactly as before                                                                                
  - The plugin registration API is unchanged: `register_handler::<H, _>` accepts both kinds                             
                                                                                                                        
  ## Test plan                            
                                                                                                                        
  - [x] `make ci` green: 213 cpex-core tests + 8 cpex-ffi tests + 51 Go tests, clippy clean, fmt clean                  
  - [x] `cargo run --example plugin_demo` produces identical scenario output (including the awaiting `RemoteAuthz`
  plugin)                                                                                                               
  - [x] `cargo run --example cmf_capabilities_demo` produces identical output                                           
  - [x] Two new tests in `manager.rs` cover the awaiting-handler path: `test_async_handler_registers_and_invokes`,
  `test_mixed_sync_and_async_handlers_in_same_hook`

### DIFF
--- a/crates/cpex-core/examples/cmf_capabilities_demo.rs
+++ b/crates/cpex-core/examples/cmf_capabilities_demo.rs
@@ -42,7 +42,7 @@ impl Plugin for IdentityChecker {
 }
 
 impl HookHandler<CmfHook> for IdentityChecker {
-    fn handle(
+    async fn handle(
         &self,
         payload: &MessagePayload,
         extensions: &Extensions,
@@ -136,7 +136,7 @@ impl Plugin for HeaderInjector {
 }
 
 impl HookHandler<CmfHook> for HeaderInjector {
-    fn handle(
+    async fn handle(
         &self,
         _payload: &MessagePayload,
         extensions: &Extensions,
@@ -201,7 +201,7 @@ impl Plugin for AuditLogger {
 }
 
 impl HookHandler<CmfHook> for AuditLogger {
-    fn handle(
+    async fn handle(
         &self,
         payload: &MessagePayload,
         extensions: &Extensions,
--- a/crates/cpex-core/examples/plugin_demo.rs
+++ b/crates/cpex-core/examples/plugin_demo.rs
@@ -76,7 +76,7 @@ impl Plugin for IdentityResolver {
 }
 
 impl HookHandler<ToolPreInvoke> for IdentityResolver {
-    fn handle(
+    async fn handle(
         &self,
         payload: &ToolInvokePayload,
         _extensions: &Extensions,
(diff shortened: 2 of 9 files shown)


araujof (approved):
LGTM; clean, well-scoped change. 

Good things: 
- The Async Fn In Trait approach
- Placement of the `+ Send` bound 
- The type-erasure composition 

All core tests + FFI tests + Go tests pass, both examples run correctly, clippy is clean.

The `RemoteAuthz` demo is a nice addition: lock drop before await, initialize-time loading, good pattern to copy.

A few suggestions for follow-up (none blocking):

1. **Doc warning on blocking calls** — now that `handle` is async, a one-liner warning that `std::thread::sleep` / blocking HTTP inside the body starves the tokio runtime (use `spawn_blocking` instead) would save someone a debugging session, especially through the FFI shared runtime path.
2. **Concurrent-mode async test** — the two new tests use Sequential mode. A variant with `PluginMode::Concurrent` would cover the full AFIT → adapter → `JoinSet::spawn` chain.
3. **Context snapshot note** — worth documenting that `&mut PluginContext` is a snapshot in concurrent mode (mutations don't propagate across handlers). More likely to bite someone now that async makes `.await`-heavy handlers natural.
4. **MSRV** — `rust-version = "1.75"` in workspace `Cargo.toml` would give downstream consumers a clean error instead of a confusing syntax failure.

PR #52, opened 2026-05-21
fix: missing cmf-demo main.go file and gitignore fix that missed it

terylt:
## Summary                                                  
                                                              
  Discovered `examples/go-demo/cmd/cmf-demo/main.go` was missing from GitHub
  even though it existed locally and `git status` didn't flag it as
  untracked. Root cause: `examples/go-demo/.gitignore` had a bare pattern                                            
  `cmf-demo` meant to ignore the built binary at the demo root, but                                                  
  gitignore matches bare patterns against **any path component** at any                                              
  depth — so `cmd/cmf-demo/` (the *source* directory) and everything                                                 
  under it were silently ignored too.                                                                                
                                                                                                                     
  Same footgun exists in the root `.gitignore` for several other common                                              
  names (`jwt/`, `build/`, `lib/`, `certs/`, `public/`, `tmp`) that                                                  
  would similarly hide future source directories. The `jwt/` one is
  particularly load-bearing — upcoming JWT-validator work needs `jwt/`                                               
  modules.                                                    
                                                                                                                     
  ## Changes                                                  
                                                                                                                     
  - `examples/go-demo/.gitignore`: anchor `cpex-demo` and `cmf-demo`                                                 
    with leading `/` so they only match the built binaries at the demo                                               
    root. Add explicit `/cmd/cmf-demo/cmf-demo` for the binary that lives                                            
    alongside the source.                                     
  - `.gitignore`: anchor `certs/`, `public/`, `build/`, `lib/` to `/`;                                               
    remove `jwt/` entirely; remove bare `tmp` (covered by `temp/`); drop                                             
    duplicate entries for `certs/`, `public/`, `build/`. Comments                                                    
    document the anchoring rationale.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,7 +29,6 @@ token.txt
 cpex.sbom.xml
 docs/docs/test/
 docs/resources/
-tmp
 *.tgz
 *.gz
 *.bz
@@ -48,8 +47,10 @@ node_modules/
 mcp.db-journal
 mcp.db-shm
 mcp.db-wal
-certs/
-jwt/
+# Anchored: matches only ./certs/ at the repo root, not nested
+# `certs/` directories under source. Bare `certs/` would silently
+# hide any cert-management module / test fixture at any depth.
+/certs/
 FIXMEs
 *.old
 logs/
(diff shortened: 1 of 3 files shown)


araujof (approved):
LGTM

PR #55, opened 2026-05-28
fix: implement __eq__ and __ne__ for CopyOnWriteDict

prakhar-singh1928:
fix: implement __eq__ and __ne__ for CopyOnWriteDict

Fixes equality comparison bug where CopyOnWriteDict compared equal to {} even when containing data from its wrapped original mapping. This caused apply_policy() to incorrectly drop valid payload modifications when plugins removed all arguments.

### Summary
CopyOnWriteDict now compares equality using its materialized logical mapping instead of the empty backing dict storage. This preserves valid hook payload updates when a plugin transforms populated args into an empty {} payload.

Closes:[ #54](https://github.com/contextforge-org/cpex/issues/54)

### Changes
- Add __eq__() to CopyOnWriteDict so equality compares logical mapping contents rather than backing storage
- Add __ne__() for consistent inequality behavior
- Explicitly mark CopyOnWriteDict as unhashable with __hash__ = None ( for lint issues)
- Add equality and regression coverage in tests/unit/cpex/framework/test_memory.py
- Add apply_policy() regression coverage in tests/unit/cpex/framework/test_policies.py
- Add executor-path regression coverage for the tool_pre_invoke empty-args flow in tests/unit/cpex/framework/test_policies.py

### Checks

- [x] PR-scoped lint issues addressed
- [ ] make lint passes
- [ ] make test passes

### DIFF
--- a/cpex/framework/memory.py
+++ b/cpex/framework/memory.py
@@ -173,6 +173,47 @@ class CopyOnWriteDict(dict):
         """
         return f"CopyOnWriteDict({dict(self.items())})"
 
+    __hash__ = None
+
+    def __eq__(self, other: Any) -> bool:
+        """
+        Compare equality with another mapping.
+
+        Compares the materialized logical mapping (original + modifications - deletions)
+        rather than the empty base dict storage.
+
+        Args:
+            other: The object to compare with.
+
+        Returns:
+            True if other is a Mapping with the same key-value pairs, False otherwise.
+            Returns NotImplemented for non-Mapping types to allow other.__eq__ to handle it.
+        """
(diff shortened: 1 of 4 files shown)


vishu-bh (approved):
Thanks @prakhar-singh1928!

Fixes CopyOnWriteDict equality to compare the materialized mapping instead of the empty underlying dict storage. That preserves valid plugin payload modifications, including the WXO case where wxo_* args are stripped and the resulting args become {}.

The added regression tests cover direct equality, policy application, and the tool_pre_invoke executor path.

LGTM ☘️

terylt (approved):
LGTM

PR #69, opened 2026-06-15
perf: shrink libcpex_ffi and harden cedar against small host stacks

araujof:
### Summary 

Size-first release profile and a leaner tokio floor cut the statically linked FFI footprint, and the cedar dispatch is made host-stack-agnostic so musl-based hosts (Alpine) stop hitting Cedar's stack guard.

### Changes 

- Add [profile.release]: opt-level="z", lto=true, codegen-units=1, strip=true. libcpex_ffi.a links statically into host binaries, so this flows into their image size; a representative statically-linked consumer shrank ~21%. panic="abort" is intentionally not set (the FFI relies on catch_unwind at its #[no_mangle] boundary). No API/ABI change.
- Trim the workspace tokio feature floor from ["full"] to ["rt","rt-multi-thread","sync","time","macros"] (the real union used by the crates); reqwest/hyper still union net/io where needed. Drops the unused fs/process/signal surface and the signal-hook-registry dep.
- Wrap the cedar dispatch (parse + build_entities + is_authorized) in apl-pdp-cedar-direct in stacker::maybe_grow. cedar-policy aborts with 'recursion limit reached' when stacker::remaining_stack() is below its 100 KiB floor; musl's 128 KiB default thread stack trips it on inputs glibc handles fine. maybe_grow runs cedar on a fresh, large segment when the host stack is low (a no-op on glibc). stacker dedups with cedar's transitive pin, so no new crates. Adds a regression test that evaluates on a 128 KiB stack.

### Tests 

770 tests pass (cargo test --workspace).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,28 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
   config visitors (it now calls `load_config_yaml` internally so `apl:`
   blocks are walked). The Go binding's `expectedFFIABIVersion` is bumped
   in lockstep.
+- Size-first `[profile.release]`: `opt-level = "z"`, `lto = true`,
+  `codegen-units = 1`, `strip = true`. `libcpex_ffi.a` is linked statically
+  into host binaries, so this flows straight into their image size — a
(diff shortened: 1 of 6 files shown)


terylt (approved):
LGTM

PR #71, opened 2026-06-17
feat: improve APL ergonomics

araujof:
### Summary

Improves APL ergonomics while maintaining backwards compatibility with original syntax. 

#### make the `apl:` wrapper optional

APL config blocks previously had to be nested under an `apl:` key at every level (`route -> apl -> policy`, `global -> apl -> pdp`, ...). The visitor's `apl_subblock` now accepts APL terms written directly on the section when the wrapper is omitted (`route -> policy`), while the explicit `apl:` form still takes precedence.

When `apl:` is absent, a synthetic block is assembled from the recognized APL keys present on the container — `policy`, `post_policy`, `args`, `result`, `pdp` — so structural keys (tool / identity / defaults / ...) are never misread. `plugins` is shape-gated: a map (the apl-override form) is included, a list (structural plugin-refs on RouteEntry / PolicyGroup) is left alone, avoiding a key-shape clash. Empty sections still return None (the "no contribution, skip" path).

Applies uniformly to the global / defaults / policy-bundle / route visit sites. Add unit tests for the shape rules (wrapper precedence, flat keys, plugins map-vs-list, null/empty) and end-to-end tests for a wrapper-less route on both the allow and deny paths.

#### accept `run` as an alias for plugin
`run(name)` now invokes a named plugin everywhere `plugin(name)` does, so policies can read `- run(audit-log)` instead of `- plugin(audit-log)`:

- policy-step / effect form (`parse_step_string`) — covers top-level  `policy:` items and `do:` lists, and `detect_step_kind` recognizes `run(` so misrouted strings still get a clear step-kind error;
- field-pipeline stage form (`args.x: "str | run(name)"`) — mirrors the  step alias for symmetry.

`plugin` remains the long form; both parse to the same `Step::Plugin` / `Stage::Plugin`. Errors name whichever verb the author used.

Add tests for the step alias (string + compile paths), the field-stage alias, and the empty/malformed error path.

#### accept unconditional `deny('reason')` as a bare action

`parse_rule`'s no-colon branch only recognized bare `deny` / `allow` via `try_bare_action`; the `deny('reason')` / `deny('reason', 'code')` call form was reachable only as the action half of a conditional rule (`predicate: deny('reason')`). That made it impossible to attach a reason/code to an unconditional reaction — notably `on_deny:` /`on_allow:` lists on PDP steps, which parse each item through `parse_step -> parse_step_string -> parse_rule`.

Fall through to `try_parse_deny_call` after `try_bare_action` so an unconditional `deny('reason')` / `deny('reason', 'code')` parses to an `Always`-guarded `Deny`. A malformed `deny(...)` surfaces its own error rather than being misread as a predicate downstream.

Add tests covering the reason-only and reason+code forms and the malformed-call error path.

### DIFF
--- a/crates/apl-core/src/parser.rs
+++ b/crates/apl-core/src/parser.rs
@@ -474,6 +474,19 @@ pub fn parse_rule(line: &str, source: &str) -> Result<Rule, ParseError> {
                     source: source.to_string(),
                 });
             }
+            // Unconditional `deny('reason')` / `deny('reason', 'code')` —
+            // the call form of a bare deny. Lets reaction lists
+            // (`on_deny: [...]` / `on_allow: [...]`) and standalone rule
+            // lines attach a reason/code without a guard predicate. A
+            // malformed `deny(...)` surfaces its own error here rather
+            // than being misread as a predicate downstream.
+            if let Some(deny) = try_parse_deny_call(trimmed, trimmed)? {
+                return Ok(Rule {
+                    condition: Expression::Always,
+                    effects: vec![deny],
+                    source: source.to_string(),
+                });
+            }
             // DSL §2 default: bare predicate denies.
             (trimmed, vec![Effect::Deny { reason: None, code: None }])
         }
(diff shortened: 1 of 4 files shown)


terylt (approved):
LGTM.  I made some small changes and you addressed the issues from my review.

PR #78, opened 2026-06-23
build: bump quinn-proto to 0.11.15 for RUSTSEC-2026-0185

araujof:
### Summary

Bump quinn-proto to 0.11.15 for RUSTSEC-2026-0185 quinn-proto < 0.11.15 is flagged by RUSTSEC-2026-0185 (remote memory exhaustion via unbounded out-of-order stream reassembly). It is pinned in the lockfile only as reqwest's optional http3 dependency and is never compiled. Bump the pin to the patched 0.11.15; no build impact.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3371,9 +3371,9 @@ dependencies = [
 
 [[package]]
 name = "quinn-proto"
-version = "0.11.14"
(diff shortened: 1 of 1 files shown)


terylt (approved):
LGTM

PR #111, opened 2026-07-01
refactor(apl)!: rename authz/authn config keys 

araujof:
## Summary

Renames the CPEX/APL authorization/authentication config keys for clarity. **Breaking change**: old key names no longer parse.

| Old | New |
|---|---|
| `identity:` | `authentication:` (global, route, policy-group) |
| `policy:` | `authorization.pre_invocation:` (or flat `pre_invocation:`) |
| `post_policy:` | `authorization.post_invocation:` (or flat `post_invocation:`) |

The two authorization phases parse equivalently whether nested under an `authorization:` block or written flat on the section.

Closes #105.

### Deliberately *not* renamed
The field-pipeline keys `args:` / `result:` stay as-is. They're aligned with the `args.*` / `result.*` attribute namespaces that predicates and interpolation read (`require(args.include_ssn)`, `${args.repo_name}`, `args.ssn | redact`) — renaming only the config block would have introduced a *new* inconsistency. (The issue originally proposed `inputs`/`outputs`; we tried it and rolled it back for exactly this reason.)

Internal APL IR (`CompiledRoute`, `Phase`) and the runtime attribute namespaces are unchanged.

## Fail-closed by design
Legacy `policy:` / `post_policy:` / `identity:` keys are **rejected loudly** at parse/load time (in both the apl-core parser and cpex-core config loader) rather than silently ignored — a dropped authorization or authentication block would otherwise fail *open*.

## Example

```yaml
routes:
  - tool: get_employee
    authentication: [jwt-user, jwt-client]
    args:
      employee_id: "str"
    authorization:
      pre_invocation:
        - "require(authenticated)"
        - "delegation.depth > 2: deny"
      post_invocation:
        - "run(audit-log)"
    result:
      ssn: "str | redact(!perm.view_ssn)"
```

## Also
- APL acronym spelled consistently as **"Authorization Policy Language"** (dropped the legacy "Attribute Policy Language").
- Docs, README, and CHANGELOG updated with the new surface and a BREAKING migration entry.

## Testing
- `cargo fmt --check`, `cargo clippy --workspace --all-targets`, and full `cargo test --workspace` (all 55 suites) pass.
- Added coverage for nested-vs-flat equivalence and legacy-key rejection.
- **Praxis regression:** built praxis against this checkout (path dep) — compiles with no API/ABI break and its cpex policy integration test passes (401/200). The `praxis-demos/demos/cpex` configs were migrated to the new surface (separate PR in that repo) and verified to load through `install_builtins` + `load_config_yaml`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,12 +15,21 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Changed
+
+- **BREAKING — APL authz/authn config keys renamed** for clarity. The old key names no longer parse; a config using them fails to load with an error naming the replacement (a dropped authorization or authentication block would otherwise fail open, so the rejection is deliberate). Migration:
+  - `identity:` → `authentication:` (at `global`, per-route, and policy-group scope)
+  - `policy:` → `authorization.pre_invocation:` (or flat `pre_invocation:`)
+  - `post_policy:` → `authorization.post_invocation:` (or flat `post_invocation:`)
+
+  The two authorization phases may be written either nested under an `authorization:` block or flat directly on the section; the forms are equivalent. The field-pipeline keys `args:` / `result:` are unchanged (they stay aligned with the `args.*` / `result.*` attribute namespaces that predicates and interpolation read). Internal APL IR is unchanged. (#105)
+
 ## [0.2.0] - 2026-06-26
 
 ### Added
 
 - CPEX redesign as a Rust framework with Go bindings
-- APL (Attribute Policy Language) governance is now bundled into `libcpex_ffi.a`. New `cpex_apl_install` extern C entry point registers the standard APL plugin/PDP factories (`validator/pii-scan`, `audit/logger`, `identity/jwt`, `delegator/oauth`, `cedar-direct`) and installs the APL config visitor on a manager. Call it after `cpex_manager_new_default` and before `cpex_load_config`. Go hosts use `PluginManager.EnableAPL()`. (#60)
+- APL (Authorization Policy Language) governance is now bundled into `libcpex_ffi.a`. New `cpex_apl_install` extern C entry point registers the standard APL plugin/PDP factories (`validator/pii-scan`, `audit/logger`, `identity/jwt`, `delegator/oauth`, `cedar-direct`) and installs the APL config visitor on a manager. Call it after `cpex_manager_new_default` and before `cpex_load_config`. Go hosts use `PluginManager.EnableAPL()`. (#60)
 - Publish `libcpex_ffi.a` as signed GitHub Release artifacts on every semver tag push (`linux-amd64-gnu`, `linux-arm64-gnu`, `linux-amd64-musl`, `linux-arm64-musl`, `darwin-arm64`). Cosign keyless signatures + SHA256 checksums; see `crates/cpex-ffi/RELEASE.md` for the schema and the verify-and-consume recipe. (#60)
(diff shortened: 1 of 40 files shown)


terylt:
Reviewed the rename mechanics end-to-end (parser, config loader, visitor, FFI/Go, docs). The core is solid and the fail-closed posture is real — a couple of things below, one of which I'd fix before merge.

## What checks out
- **Legacy rejection is thorough.** `policy:` / `post_policy:` are rejected at every scope in both the apl-core parser (`reject_legacy_keys` in `compile_route` + `compile_apl_blocks`) and the cpex-core visitor (`reject_legacy_apl_keys` at global / default / policy-group / route). `identity:` is rejected at global, `global.policies.*`, `global.defaults.*`, and each route in `reject_renamed_identity_key`, which is actually called (config.rs:628) before validation. No scope silently drops a legacy key at the top level.
- **FFI / Go bindings** carry no stale config keys — the only `identity` left is the plugin *kind* `identity/jwt`, which is correctly untouched.
- **Docs** are fully converted — no lingering `policy:` / `post_policy:` / `identity:` in any example that would now hard-fail to load.
- **Keeping `args:` / `result:` as-is** is the right call — renaming only the config block while the `args.*` / `result.*` attribute namespaces stay would have traded one inconsistency for another.

## Would fix before merge

**Fail-open: a legacy key nested inside the new `authorization:` block is silently dropped.**
`AuthorizationYaml` (parser.rs) has only `pre_invocation` / `post_invocation` — no `#[serde(flatten)]` capture and no `deny_unknown_fields`. So this:

```yaml
routes:
  - tool: get_employee
    authorization:
      policy:            # <- legacy key, nested under the new wrapper
        - "require(authenticated)"
```

deserializes with `pre_invocation` empty, the inner `policy` silently ignored, and **no error** — the route loads with no authorization enforced. `reject_legacy_keys` doesn't catch it because it inspects `RouteYaml.other`, but `policy` here was consumed as part of the `authorization` value, so it never lands in `other`.

This is a plausible partial-migration mistake: reading the migration entry `policy: → authorization.pre_invocation:`, it's easy to write `authorization:` and then habitually keep `policy:` underneath. It's the exact fail-open the rest of the PR guards against, just at one nesting the guard doesn't reach.

Fix: add `#[serde(deny_unknown_fields)]` to `AuthorizationYaml` (it has no `flatten`, so this is safe, and it also catches typos like `pre_invocaton:`). A regression test for `authorization: { policy: [...] }` → load error would lock it in.

## Minor

**Both forms on the same section silently merge (double-execution footgun).** `compile_apl_blocks` does `auth_pre.iter().chain(raw.pre_invocation.iter())`, so if a section declares *both* nested `authorization.pre_invocation` and flat `pre_invocation`, the entries concatenate rather than conflict. For idempotent denies that's harmless, but a `run(audit-log)`, a Transform/mutating plugin, or a `delegate()` declared in both forms runs **twice**. Given the PR fails loud on almost everything else, silently merging both-on-the-same-section is a little inconsistent — worth either a doc note or rejecting the both-present case. (Mixing forms across *different* scopes — e.g. global nested, route flat — is fine; they're meant to stack.)

## Naming (non-blocking)

Overall I'm fine with the direction — `identity: → authentication:` is a clear win, and `pre_invocation` / `post_invocation` fixes the old `policy` / `post_policy` asymmetry cleanly. One small semantic note for the record: the pre-invocation phase isn't a binary allow/deny — it compiles to effects that include `Elicit` (human approval / suspend), `Taint`, `Delegate`, and `Plugin` (which can `modify_payload`), so "authorization" slightly undersells it. The nice thing is the flat `pre_invocation:` / `post_invocation:` form sidesteps that entirely (it names *when*, not *what*), so anyone who finds `authorization:` too narrow can just use the flat keys. Since the wrapper is optional, this isn't something to block on — just worth a sentence in the docs that the phase runs a decision *plus* obligations/effects, so nobody reads `authorization:` as a pure gate.

